PR #19, opened 2015-03-15
Fix types

emmt:
* fix integer types (`Int32` --> `Cint`) and double check the prototypes in `fitsio.h` header
* fix wrong type in `fits_get_rowsize/ffgrsz`
* `ffhdr2str` returns a `Cint` not a `Ptr{Ptr{Uint8}}`
* `fits_get_img_size/ffgisz` and `fits_create_img/ffcrim`: argument `naxes` is an array of long's
* use `ffcrimll` for `fits_create_img` and `ffppxll` for `fits_write_pix` instead of `ffcrim` and `ffppx` to let the user call these methods with any vector of integers (and save him/her the bothering of doing any conversion)
* in `fits_read_col` and `fits_write_col` no needs to convert to `Int64`, this is automatically done by `ccall`
* enlarge the accepted types for arguments of the methods (e.g. `keynum::Int` --> `keynum::Integer` in fits_read_record), `ccall` does the conversion
* use `@eval` macro to create data type tables and related methods, also added the table summarizing the assumptions about what types should be used (for future reference, maybe it can be in the documentation instead) missing table types 40 and 41 were added (for C `unsigned long` and C `long`) for now some methods may override another one (but this is is harmful, can be avoided later when building the dictionary of FITS table type code to Julia types which is needed to read table columns)

### DIFF
--- a/src/cfitsio.jl
+++ b/src/cfitsio.jl
@@ -1,5 +1,109 @@
 using Compat
 
+# functions to convert to C integer types (this could be in Julia Base module):
+for (f, T) in ((:cint,  Cint),
+               (:clong, Clong))
+    @eval begin
+        ($f)(x::$T) = x
+        ($f)(x::Integer) = convert($T, x)
+        ($f)(x::AbstractArray{$T}) = x
+        function ($f){T<:Integer}(x::AbstractArray{T})
+            y = similar(x, $T)
+            i = 1
+            for e in x
+                y[i] = convert($T, e)
+                i += 1
+            end
+            y
+        end
+    end
+end
+
+# The following table gives the correspondances between CFITSIO "types",
+# the BITPIX keyword and Julia types.
+#
+#     -------------------------------------------------
+#     CODE  CFISTIO         Julia     Comments
+#     -------------------------------------------------
+#           int             Cint
+#           long            Clong
+#           LONGLONG        Int64     64-bit integer
+#     -------------------------------------------------
+#     -------- FITS BITPIX ----------------------------
+#        8  BYTE_IMG        Uint8
+#       16  SHORT_IMG       Int16
+#       32  LONG_IMG        Int32
+#       64  LONGLONG_IMG    Int64
+#      -32  FLOAT_IMG       Float32
+#      -64  DOUBLE_IMG      Float64
+#      -------- cfitsio "aliases" ---------------------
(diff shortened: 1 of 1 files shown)


src/cfitsio.jl
```diff
@@ -502,12 +553,8 @@ function fits_read_col{T}(f::FITSFile,
                           firstelem::Integer,
                           data::Array{T})
 
-    anynull = Int32[0]
-    status = Int32[0]
-
-    firstrow64 = convert(Int64, firstrow)
-    firstelem64 = convert(Int64, firstelem)
-    nelements64 = convert(Int64, length(data))
```
kbarbary: Should there be a line `nelements = length(data)` here? (`nelements` is used below in ccall)

src/cfitsio.jl
```diff
@@ -1,5 +1,109 @@
 using Compat
 
+# functions to convert to C integer types (this could be in Julia Base module):
```
kbarbary: These conversion functions are fine with me for now, but note that this wouldn't be accepted into Julia Base because conversion functions have been changed in Julia v0.4. Functions like `int32()` have been removed in favor of `Int32(x)` for converting scalars and `Array{Int32}(x)` for converting arrays.

(Eventually we should switch to the new syntax and use Compat to keep supporting v0.3, but that's a different PR!)

kbarbary:
Other than my comment about a missing `nelements` this all looks good to merge.

src/cfitsio.jl
emmt: Le 15/03/2015 23:02, Kyle Barbary a écrit :
> In src/cfitsio.jl
> <https://github.com/JuliaAstro/FITSIO.jl/pull/19#discussion_r26452973>:
>
>> @@ -502,12 +553,8 @@ function fits_read_col{T}(f::FITSFile,
>>                            firstelem::Integer,
>>                            data::Array{T})
>>
>> -    anynull = Int32[0]
>> -    status = Int32[0]
>> -
>> -    firstrow64 = convert(Int64, firstrow)
>> -    firstelem64 = convert(Int64, firstelem)
>> -    nelements64 = convert(Int64, length(data))
>
> Should there be a line |nelements = length(data)| here? (|nelements| is
> used below in ccall)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/JuliaAstro/FITSIO.jl/pull/19/files#r26452973>.
>


Yes you are right, I introduced the same error in fits_write_col
these are now fixed

PR #20, opened 2015-03-17
Pull request/1a150c30

emmt:
Same changes as before plus fix of unitialized variable `nelements` in `fits_read_col` and `fits_write_col`.

### DIFF
--- a/src/cfitsio.jl
+++ b/src/cfitsio.jl
@@ -1,5 +1,109 @@
 using Compat
 
+# functions to convert to C integer types (this could be in Julia Base module):
+for (f, T) in ((:cint,  Cint),
+               (:clong, Clong))
+    @eval begin
+        ($f)(x::$T) = x
+        ($f)(x::Integer) = convert($T, x)
+        ($f)(x::AbstractArray{$T}) = x
+        function ($f){T<:Integer}(x::AbstractArray{T})
+            y = similar(x, $T)
+            i = 1
+            for e in x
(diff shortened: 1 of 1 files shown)


kbarbary:
Looks good. For future reference, you can add commits to an existing PR, rather than creating a new one. Just add the commits to the same branch and push them to your github account and they will automatically show up in the PR. This is easier if you start by creating a branch (using command-line git) that has a more memorable name, like `fix-types` in this case. That way you can be sure you're adding commits to the correct branch. The [latest manual](http://julia.readthedocs.org/en/latest/manual/packages/#code-changes) has fairly good tips, but personally I tend to just use command-line git directly for most operations rather than `Pkg` from within Julia.

For this case, I'll merge this PR and close #19 as "superceded."

kbarbary:
Travis reports a test failure on Julia 0.4, but reports that #19 passed on Julia 0.4. I don't see how the additional (2-line) commit would have caused a test failure, so basically I'm going to ignore it.

emmt:
OK thanks for these guidelines, I agree that it is better the way you 
suggest.

emmt:
What are these mails that I receive from 'coveralls' ?

I hope I didn't do a mistake, but in principle there are more 
capabilities in the last commit for FITSIO...

kbarbary:
Coveralls is a service that reports changes in the test coverage (% of code lines run in the test suite). The comments are a bit annoying though... I'll try to turn them off.

PR #22, opened 2015-03-18
emit warnings for deprecated functions

kbarbary:
- move the definitions of `fits_get_coltype`, `fits_get_eqcoltype`, `fits_read_tdim` to earlier in the file so that `fits_get_coltype` can be used in `fits_read_col` instead of the deprecated function `fits_get_col_repeat`.
- move deprecated functions/methods to a separate file `deprecations.jl` and use the `@deprecated` macro to emit a deprecation warning.

@emmt should review this. Tests pass locally but there's no tests for the table functions yet, so it is not a very good indicator!

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,0 +1,16 @@
+v0.6.0 (unreleased)
+===================
+
+## Deprecations in low-level c-style interface
+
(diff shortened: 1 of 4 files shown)


kbarbary:
I guess there's no commentary on this or #23 so I'll just merge both from #23.

Issue #24, opened 2015-03-27
Problem with FITSIO high level interface


sherrinm:
I have had problems with reading the header of  an M31 Andromeda FITS file - which was part of my wife's S177 UK Open University Astronomy project - using the HIGH level routines

The file is from the http://www.spacetelescope.org  (NASA/ESA Hubble Space Telescope) so has reasonably parentage :

http://www.spacetelescope.org/static/projects/fits_liberator/datasets/m31/f001a066.zip

using FITSIO
cd ("S177")

f001a = FITS("f001a066.fits")
file: f001a066.fits
mode: r
extnum exttype         extname
1      image_hdu  

size(f001a[1])
(7055,7055)

data = read(f001a[1]); 
data[7055,7055]
29976

header = readheader(f001a[1])
ERROR: couldn't parse keyword value: "1.9511305786508E+00,"
 in parse_header_val at /Users/malcolm/.julia/v0.3/FITSIO/src/hdutypes.jl:161
 in readheader at /Users/malcolm/.julia/v0.3/FITSIO/src/hdutypes.jl:204

The keyword value has a trailing ',' which throws the error in parse_header_val

Using the LOW level API gave no problems:
 
f001 = fits_open_file("f001a066.fits")
n = fits_get_hdrspace(f001)[1]
for i = 1:n
   println(fits_read_keyn(f001,i))
end

Looking at the header from a LOW level data dump, the problem is with the SKEW value but I suspect it is actually the entry above which has no key or value, just a comment, which I suspect is the real culprit

("CROTA1","1.9373114911858E+00","Rotation angle of first axis (deg)")
("CROTA2","1.9373114911858E+00","Rotation angle of second axis (deg)")
("","","Warning: CROTA2 is inaccurate due to considerable skew")
("SKEW","1.9511305786508E+00,","1.9234924037208E+00 /Measure of skew") 

A work around in the parse_header_val routine  (my hack-- I am a reformed Perl programmer after all)  was : 

function parse_header_val(my_val::String)
    if length(my_val) == 0 return "" end
    val = (my_val[end]==',')  ?  chop(my_val)  : my _val
    try
        return int(val)  
etc....

But clearly a fix is better in reader header() to ensure that the values don't have trailing ',' or other nasty characters to begin with.

kbarbary:
I don't think the warning comment above is a problem here.

Looking at the raw header,
```
[...clip...]
[...]

PR #25, opened 2015-03-27
non-standard header values

kbarbary:
This is a simple fix for #24. To explain: cfitsio returns header values as strings. In the high-level interface `FITSHeader`, we parse these strings into their implied types, for convenience. First we try to parse as an int, then a float, then a bool, then nothing, then a string (if value is surrounded by `'`). What to do when we get to the end? Formerly we raised an error saying that the value couldn't be parsed. This PR changes the behavior to simply return the unparsed string.

The upside of the change is that a single malformed keyword will no longer prevent one from reading the entire header. The slight downside is that you can't tell if the original unparsed value was a correctly quoted string or a malformed value that we gave up parsing. The upside outweighs the downside, but it made me think about whether we should keep the unparsed header in memory so that we can write it back out *exactly* as the original.

~~While checking the behavior of this non-standard file I also ran into another bug that I'll add a fix for here before we merge.~~ *edit:* Done. Formerly, records that were not COMMENT or HISTORY with an empty value field caused writing the header out to fail, even though these are valid FITS. I added a method to `fits_update_key()` for `Nothing`/`Void` values.

*edit:* Added regression test for such non-standard keywords like the ones in #24. We can add future pathological cases along side these.

*edit:* Fixed a problem with `show(::IO, ::FITSHeader)` where lines could wrap past 80 characters (became apparent with a comment that was longer than 47 characters).

### DIFF
--- a/src/cfitsio.jl
+++ b/src/cfitsio.jl
@@ -14,7 +14,8 @@
 #   `@compat Vector{T}(x)` once syntax support is in Compat.
 # - `convert(Cint, x)` can be changed to `Cint(x)` once v0.3 is not supported
 #   or `@compat Cint(x)` once syntax support is in Compat.
-#
+# - Once v0.3 is no longer supported, the type `Nothing` should be changed to
+#   `Void` (this is the type of `nothing`).
 #
 # The following table gives the correspondances between CFITSIO "types",
 # the BITPIX keyword and Julia types.
@@ -299,6 +300,15 @@ function fits_update_key(f::FITSFile, key::ASCIIString, value::FloatingPoint,
     ccall(("ffukyd", libcfitsio), Cint,
           (Ptr{Void}, Ptr{Uint8}, Cdouble, Cint, Ptr{Uint8}, Ptr{Cint}),
           f.ptr, key, value, -15, comment, status)
+    fits_assert_ok(status[1])
+end
+
+function fits_update_key(f::FITSFile, key::ASCIIString, value::Nothing,
+                         comment::Union(ASCIIString, Ptr{Void})=C_NULL)
+    status = Cint[0]
+    ccall(("ffukyu", libcfitsio), Cint,
+          (Ptr{Void}, Ptr{Uint8}, Ptr{Uint8}, Ptr{Cint}),
+          f.ptr, key, comment, status)
     fits_assert_ok(status[1])
 end
 
--- a/src/hdutypes.jl
+++ b/src/hdutypes.jl
@@ -157,11 +157,19 @@ function parse_header_val(val::ASCIIString)
         catch
         end
     end
-    val == "T" ? true :
-    val == "F" ? false :
-    val == "" ? nothing :
-    length(val) > 1 && val[1] == '\'' && val[end] == '\'' ? val[2:end-1] :
-    error("couldn't parse keyword value: \"$val\"")
+    if val == "T"
+        return true
+    elseif val == "F"
+        return false
+    elseif val == ""
+        return nothing  # The value area is empty.
+    elseif length(val) > 1 && val[1] == '\'' && val[end] == '\''
+        return val[2:end-1]  # The value is a string. Strip the quotes.
+                             # TODO: strip any trailing white space?
+    else
+        return val  # The value (probably) doesn't comply with the FITS
+                    # standard. Give up and return the unparsed string.
+    end
 end
 
 function readkey(hdu::HDU, key::Integer)
--- a/src/hdutypes.jl
+++ b/src/hdutypes.jl
@@ -165,7 +165,6 @@ function parse_header_val(val::ASCIIString)
         return nothing  # The value area is empty.
     elseif length(val) > 1 && val[1] == '\'' && val[end] == '\''
         return val[2:end-1]  # The value is a string. Strip the quotes.
-                             # TODO: strip any trailing white space?
     else
         return val  # The value (probably) doesn't comply with the FITS
                     # standard. Give up and return the unparsed string.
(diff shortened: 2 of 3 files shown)


kbarbary:
I disagree that we should emit a warning. Warnings can be useful when they're something that a user can fix (deprecation warnings, for example). In this case, the user almost certainly has no control over the format of the file they're reading, so she or he can't do anything to make the warning go away. From experience with some FITS parsers, this can be quite annoying. Furthermore, there might only be one keyword record that was non-standard; it's quite possible (probable, even) that the user isn't concerned with the value of that specific record, and that the rest of the header parsed just fine.

If it helps, I don't think of the new behavior as "ignoring an error." Rather, we are redefining the parsing behavior as follows:

- starting from the value field in string form, attempt to parse it according to the FITS standard
- if the value is non-standards compliant, return the unparsed value string

I think simply clearly documenting this behavior should be enough (and I can add this to the docs).

sherrinm:
Doesn't this solution mean the the HIGH level and LOW level APIs may return different values in certain cases.

As such would I be tempted not to use the HIGH level API?

kbarbary:
@sherrinm I'm not sure I understand what you mean. Even before this change, the high-level interface returned parsed values and the low-level interface returned unparsed strings, so the returned values have always been different.

sherrinm:
I guess my point is that it is possible to split the three fields on each line via the ',' separator and parse the floating point values. Returning a float with a trailing comma is not quite the same thing.

kbarbary:
I feel like there's a little confusion here, because at no point is there a string with three fields separated by a comma. Just so we're on the same page, the raw header on disk is a string of ASCII characters, an excerpt of which looks like this:

```
SKEW    =  1.9511305786508E+00,  1.9234924037208E+00 /Measure of skew           
CDELT1  = -2.8324823369599E-04 /RA pixel step (deg)                             
```

cfitsio's `fits_read_keyn` parses a given header line into three strings:

```julia
julia> fits_read_keyn(f, 119)
("CDELT1","-2.8324823369599E-04","RA pixel step (deg)")
```
Note that this is a tuple of three strings - the comma is only there for the viewer's sake. You can also do:

```julia
julia> key, value, comment = fits_read_keyn(f, 119);

julia> key
"CDELT1"

julia> value
"-2.8324823369599E-04"

julia> comment
"RA pixel step (deg)"
```
The high level interface simply takes the extra step of converting the value (originally an ASCIIString `"-2.8324823369599E-04"` into the Float64 value `-0.00028324823369599`.

Now, what happens with the problematic line reported in #24 and reproduced in the raw header snippet above? The raw header looks like this:

```
SKEW    =  1.9511305786508E+00,  1.9234924037208E+00 /Measure of skew           
```

Parsing using the low level interface:
```julia
julia> key, value, comment = fits_read_keyn(f, 118);

julia> key
"SKEW"

julia> value
"1.9511305786508E+00,"

julia> comment
"1.9234924037208E+00 /Measure of skew"
```

Again, in the high-level interface, we'd like to further parse the value string `"1.9511305786508E+00,"` into the intended type - one of ASCIIString, Float64, Int, Void or Bool. However, the format of the string doesn't follow any of the FITS conventions for any of those types, so the intended type is ambiguous. In this PR, we simply leave the value as the string `"1.9511305786508E+00,"`. In this case the value returned by the low level API (using `fits_read_keyn`) and the high-level interface will in fact be identical. It's only in cases where we successfully parse the value into the intended type that they differ.

sherrinm:
Don't think I'm confused

I'm not saying its a type with 3 string fields, just that it is easy to
split on the comma.
But in the end its your call.

kbarbary:
@sherrinm Sorry for the over-explanation then. Unfortunately, I still don't understand what you mean by "split on the comma", so I guess I'm the confused one.

sherrinm:
No sweat.
I've been confused lots of times in the past, so might well be again.
I need to look more at the M31 FITS file.

For parsing the Low level header strings seems to me its possible to use
regular expressions (I'm a reformed Perl munger after all) or dumping to a
string with fixed 80-byte records using *fits_hdr2str()* and slicing it up,
i.e key [1:8], value[10:31], comment[33:80]

I'm in Darmstadt for the rest of the week, so I'll have a fiddle with the
'rogue' dataset when I return to the UK.

I like the convenience of the High level API but if a value is "supposed"
to be a number I want to be able to use it, or at least say float(val)
without catching conversion errors -- if that makes sense?

kbarbary:
Thaks for explaining more. The function `fits_read_keyn()` is already doing the slicing you suggested. For the original header line:
```
SKEW    =  1.9511305786508E+00,  1.9234924037208E+00 /Measure of skew           
```

If you do

```julia
julia> key, value, comment = fits_read_keyn(f, 118);
```
the key is characters `1:8`, the value is characters `11:30` and the comment is the rest of the line. `fits_read_keyn` doesn't *just* do splitting though. It takes care of stripping trailing whitespace, and it deals with string values that extend past the normal value area (which is allowed by the standard).

That leaves us with the issue of how to convert the value string `"1.9511305786508E+00,"` to a float. Yes, we could put in code to strip trailing commas and then do a conversion. However, while that may solve this one particular case, there's no guarantee that it will solve any other case. What if there had been two fewer significant digits, so that the value string was `"1.95113057865E+00, 1"`? What if a colon had been used as the separator? Without assuming the header follows some standard, there are a million different scenarios and we can't possibly code for all of them. Even if we could, we wouldn't get the *intended* value in this case, which is actually the two-float value: `1.9511305786508E+00,  1.9234924037208E+00`.

Most of the time, the header values will be standard-compliant and the high-level interface will just work, returning values as the intended types.  In non-standard cases like this one, you can still do a little custom parsing to recover the value if needed.

Finally, It's instructive to see what other libraries do.

**fitsio** (Python)

```python
>>> f = fitsio.FITS("f001a066.fits")

>>> hdr = f[0].read_header()

>>> hdr["SKEW"]
(1.9511305786508,)  # just the first value, parsed into a Python tuple
```

**astropy.io.fits**
```python
>>> from astropy.io import fits
>>> f = fits.open("f001a066.fits")
>>> hdr = f[0].header
>>> hdr["SKEW"]  # initially errors...
astropy.io.fits.verify.VerifyError: Unparsable card (SKEW), fix it first with .verify('fix').
>>> f.verify('fix')
WARNING: VerifyWarning: Verification reported errors: [astropy.io.fits.verify]
WARNING: VerifyWarning: HDU 0: [astropy.io.fits.verify]
WARNING: VerifyWarning:     Card 117: [astropy.io.fits.verify]
WARNING: VerifyWarning:         Card 'SKEW' is not FITS standard (invalid value string: '1.9511305786508E+00,  1.9234924037208E+00 /Measure of skew').  Fixed 'SKEW' card to meet the FITS standard. [astropy.io.fits.verify]
WARNING: VerifyWarning: Note: PyFITS uses zero-based indexing.
 [astropy.io.fits.verify]
>>> hdr["SKEW"]
'1.9511305786508E+00,  1.9234924037208E+00'  # after "fixing" returns a string with both values
```

The astropy.io.fits behavior is actually not too bad here. It errors only when the user tries to access the specific card, and then provides a way to fix it. That sort of behavior is beyond the scope of this PR though.

kbarbary:
I'm going to go ahead and merge this. It's a simple change that is a strict improvement over the current situation. That's not to say its the final word on parsing non-standard header records! But let's discuss further modifications in separate issues.

PR #28, opened 2015-04-25
explicitly use Tuple with typealias ColumnDef

nirinA:
this quick fix prevents error in ColumnDef with recent v0.4

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,3 @@
 julia 0.2
 BinDeps
-Compat 0.3.4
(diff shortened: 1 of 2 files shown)


kbarbary:
I think this should be `Tuple{...}` not `Tuple(...)`. Also should bump the required version of Compat to v0.4.1 in REQUIRE.

nirinA:
ouch! corrected and bumped
thanks

kbarbary:
Thanks!

Issue #27, opened 2015-04-08
Discourage and eventually deprecate "low-level" API


kbarbary:
In the long term, I think it will become a nightmare having the dual interface that currently exists in this package. Instead, we should ensure that the high-level interface has all the functionality *and performance* possible in the low-level interface and eventually deprecate the low-level interface for public use. (I've been putting off having this discussion because it might be contentious, but it will be better to resolve sooner rather than later!)

Obviously there are still gaps in the high-level interface, the most glaring of which is support for tables. I'm currently working on a PR to support tables in the high-level interface (at least at the level available in the low-level interface). The other thing I can think of off the top of my head is a way to get "raw" header values as strings.

What else is missing from the high-level interface, in terms of either functionality or performance? If you're currently using the low-level API, could you switch?

emmt:
I agree and I have avoided mixing low and high level interface.  The current OIFITS has been fixed to cope with that.

ziotom78:
As a matter of fact, I never used FITSIO.jl's high-level interface as I work with tables 99% of time. I usually prefer working with high-level interfaces, but the case of CFITSIO is somewhat difference in my opinion. The advantage of the low-level interface is its being a 1-to-1 mapping to the C/Fortran functions provided by CFITSIO, which are widely used and well known. It is extremely easy for me to port some FITS-based code from/to Julia if this mapping still exists.

To provide a concrete example, I find [pyfits](http://www.stsci.edu/institute/software_hardware/pyfits) annoying to use, because of the very abstractions it provides. The few times I have used it, I have always had to dig into the documentation to know how to do things. Had it provided a low-level interface, I would have probably knew by hearth which function to use and what parameter it expected.

kbarbary:
@ziotom78 I agree about pyfits/astropy.io.fits being "too-abstracted". In the FITSIO.jl high-level interface, I've been mimicing the fitsio python package, which I think is less abstracted and makes it easier to understand what is happening. For example, I'm interested in keeping it very clear when read/write operations actually happen, which is something I find difficult to understand in astropy.io.fits.

(As an aside that is related to this, I'm thinking about deprecating the use of `hdu[i:j]` as a way to read image subsections, and replacing it with `read(hdu, i:j)`, again to make clear that there is a read operation happening. The current behavior is a bit too cute.)

[...]

PR #30, opened 2015-05-04
Rename a few functions, create Libcfitsio submodule

kbarbary:
This PR consists of two commits:

## Renames

The first commit renames some functions in the high-level API according to the style guide:

- `readkey` -> `read_key`
- `readheader` -> `read_header`
- `getcomment` -> `get_comment`
- `setcomment!` -> `set_comment!`

It's my fault that these functions weren't named this way to begin with. I was trying to mimic naming conventions in Base (e.g., `haskey(d)`), not realizing that the style guide is a "do as we say, not as we do" sort of thing.

It also deprecates the `hdu[i:j, k:l]` syntax for reading a subset of an image HDU in favor of the more explicit `read(hdu, i:j, k:l)`. I really like this change as it makes all read and writes from the file explicit, making the whole API more consistent.

There are deprecation warnings for all these changes.

## Libcfitsio submodule

The second commit moves all of the low-level interface into a `Libcfitsio` submodule, as proposed in #27. This would require users of the low-level `fits_*` functions to do `using FITSIO.Libcfitsio` rather than just `using FITSIO`. Otherwise all low-level code would remain the same. What do users of the low-level API think about this? cc @emmt @ziotom78 

There are **not** deprecation warnings for this change, but users only need to replace `using FITSIO` with `using FITSIO.Libcfitsio` to comply with the change.


ziotom78:
@kbarbary , thanks a lot for this PR, as I said before I find your idea of having a low-level submodule perfect. No objections about changing `using FITSIO` into `using FITSIO.Libcfitsio`, it's a minor change that's easy to remember.

emmt:
What you propose is fine for me. I will modify `OIFITS.jl` to reflect these as soon as possible. Do you plan to make a new release of `FITSIO.jl`?

kbarbary:
@emmt Yes, after this is merged would be a good time for a new release.

Issue #33, opened 2015-05-13
Key not found when reading from binary table


rgiordan:
I'm having trouble reading a column from a binary table.  The file is from [this URL](http://data.sdss3.org/sas/dr10/boss/photo/redux/301/3900/objcs/6/psField-003900-6-0269.fit).  I'm new to FITS, so apologies in advance if I'm missing something basic.

I would like to read the RROWS column.  It is present:

```
psf_filename = "psField-003900-6-0269.fit"
psf_fits = FITS(psf_filename);
read_header(psf_fits[b + 1])
# ...other fields...
# TFORM7  =                 '1J' / PIXDATATYPE
# TTYPE7  =              'RTYPE' / type
# TFORM8  =             '1PE(0)' / FLOAT
# TTYPE8  =              'RROWS' / rows_fl32
# TFORM9  =                 '1J' / INT
# TTYPE9  =              'RROW0' / row0
# TFORM10 =                 '1J' / INT
# TTYPE10 =              'RCOL0' / col0
# ....
```

And I can read other fields:

```
read(psf_fits[b + 1], "RNROW")
# 4-element Array{Int32,1}:
#  51
#  51
#  51
#  51
```

But I cannot read RROWS:

```
read(psf_fits[b + 1], "RROWS")
# ERROR: key not found: -42
#  in read at /home/rgiordan/.julia/v0.3/FITSIO/src/hdutypes.jl:72
```

Any idea what might be going wrong?

kbarbary:
Thanks for reporting.

It looks like this column of the table holds a "variable length array". Unfortunately, support for variable length arrays has not yet been implemented in FITSIO.jl. But this is good motivation to do so! I'll try to look into it soon.

rgiordan:
Thanks for the quick response.  Is there a low-level workaround in the meantime?

kbarbary:
I pushed a very small change to the `var-arrays` branch on my fork that allows reading variable length arrays in the high-level interface. The commit is here: https://github.com/kbarbary/FITSIO.jl/commit/4275faef4e5757ca7589ace9860e9e16342c62ce

However, I'd like to test this before merging into master to make sure it's correct. If you could test it out, that would also be helpful.

rgiordan:
Yeah, it no longer gives a missing column error, but I get four zeros:
```
read(psf_fits[b + 1], "RROWS")
# 4-element Array{Float32,1}:
#  0.0
#  0.0
#  0.0
#  0.0
[...]

PR #36, opened 2015-05-21
Read/write variable-length table columns

kbarbary:
On reading, variable-length columns are automatically read into `Vector{Vector{T}}` or `Vector{ASCIIString}` (for string columns).

On writing a new table, the `varcols` keyword can be used to specify which columns should be written as variable length. This is necessary because `Vector{ASCIIString}` can go either way. Other types must be vectors of vectors in order to be written as variable-length.


kbarbary:
Also: addresses #33.

Issue #39, opened 2015-06-22
Raw header for WCSLIB


rgiordan:
I'm looking into getting WCS information out of FITS image files, and it seems like the preferred function to use is [```wcspih```](http://www.atnf.csiro.au/people/mcalabre/WCS/wcslib/wcshdr_8h.html#ac75623ee805ab7d43b0bba684c719a60), which, in [```WCSLIB.jl```](https://github.com/JuliaAstro/WCSLIB.jl/blob/master/src/WCSLIB.jl) at least, requires a raw ASCII FITS header.  I think this can only be accessed with the now private ```Libcfitsio.fits_hdr2str``` function.

I think this could also be done with ```wcsprm```, but I haven't been able to find clear enough documentation about exactly which header key values should go into which arguments to do this with confidence.

Do you have thoughts about whether there is a better way to do this, or whether maybe the raw header string should be available to users for this reason?

kbarbary:
I think it does indeed make sense to expose this FITSIO functionality in the high-level interface. How about adding a method `read_header(hdu::HDU, ::Type{ASCIIString})` so that, e.g., 

```julia
raw_header = read_header(f[1], ASCIIString)
```

returns the header as a single string, rather than a `FITSHeader` object?

Ideally, in the future a `wcsprm` object could be constructed by passing any mapping-like object (e.g., `Dict` or `FITSHeader`) that contains WCS keywords. That's obviously a bit more work though.

rgiordan:
Sounds good, let me try to put together a pull request.

PR #40, opened 2015-06-22
Read raw header from an HDU

rgiordan:
For [Issue 39](https://github.com/JuliaAstro/FITSIO.jl/issues/39).


kbarbary:
Added news entry, docstring and a test.

PR #41, opened 2015-06-23
read_header raw string bug fix

rgiordan:
Hey Kyle,

I'm sorry, there was a bug in my read_header.  I made a fix and added a test.

Ryan


kbarbary:
Oh yeah, I forgot about that!

PR #45, opened 2015-09-29
Fixes Julia 0.4-rc2 warnings 

jeff-regier:
Hi Kyle, This PR fixes all the deprecated warnings from FITSIO.jl caused by the recent changes to Julia 0.4.

### DIFF
--- a/src/FITSIO.jl
+++ b/src/FITSIO.jl
@@ -74,11 +74,11 @@ end
 # in the HDU object.
 type FITS
     fitsfile::FITSFile
-    filename::String
-    mode::String
+    filename::AbstractString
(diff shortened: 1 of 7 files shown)


kbarbary:
Nice! One question: are the `@compat(UInt8)` necessary, or can you just do `UInt8` once `Compat` is used? The later seems to work for me in the REPL.

jeff-regier:
Oh, I see...it works for me too on the command line. I guess I shouldn't
have put @compat everywhere.

jeff-regier:
I've now removed the unnecessary calls to `@compat`.

kbarbary:
Thanks!

PR #48, opened 2016-02-29
Enable precompilation

kbarbary:


kbarbary:
Failure on Julia nightly 64-bit windows is from Julia itself.

PR #53, opened 2016-07-28
fix warnings from Julia 0.5 

jeff-regier:
Hi @kbarbary  -- This fixes the vast majority of warnings from Julia 0.5, without breaking any backward compatibility, I think.

### DIFF
--- a/src/header.jl
+++ b/src/header.jl
@@ -6,7 +6,7 @@
 # Used here and in other files. Functions that operate on FITSFile
 # start with `fits_`.
 
-function try_parse_hdrval(::Type{Bool}, s::ASCIIString)
+function try_parse_hdrval(::Type{Bool}, s::Compat.ASCIIString)
     if length(s) == 1
         if s[1] == 'T'
             return Nullable(true)
@@ -22,9 +22,9 @@ end
 # single space).  See CFITSIO manual section 4.5 for details.
 #
 # TODO: parse '' within the string as a single '.
-function try_parse_hdrval(::Type{ASCIIString}, s::ASCIIString)
+function try_parse_hdrval(::Type{Compat.ASCIIString}, s::Compat.ASCIIString)
     if length(s) < 2 || s[1] != '\'' || s[end] != '\''
-        return Nullable{ASCIIString}()
+        return Nullable{Compat.ASCIIString}()
     end
 
     i = endof(s) - 1
@@ -37,13 +37,13 @@ function try_parse_hdrval(::Type{ASCIIString}, s::ASCIIString)
     return Nullable(s[2:i])
 end
 
-try_parse_hdrval(::Type{Float64}, s::ASCIIString) = tryparse(Float64, s)
+try_parse_hdrval(::Type{Float64}, s::Compat.ASCIIString) = tryparse(Float64, s)
 
 # hack for integers in Julia v0.3: tryparse(Int, s) not available in Compat.
 if VERSION > v"0.4.0-dev+3864"
-    try_parse_hdrval(::Type{Int}, s::ASCIIString) = tryparse(Int, s)
+    try_parse_hdrval(::Type{Int}, s::Compat.ASCIIString) = tryparse(Int, s)
 else
-    try_parse_hdrval(::Type{Int}, s::ASCIIString) = try
+    try_parse_hdrval(::Type{Int}, s::Compat.ASCIIString) = try
         Nullable(parseint(s))
     catch e
         Nullable{Int}()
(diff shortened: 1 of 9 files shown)


src/libcfitsio.jl
```diff
@@ -208,7 +208,7 @@ function fits_create_file(filename::AbstractString)
     ptr = Array(Ptr{Void}, 1)
     status = Cint[0]
     ccall((:ffinit,libcfitsio), Cint, (Ptr{Ptr{Void}},Ptr{UInt8},Ptr{Cint}),
-          ptr, bytestring(filename), status)
+          ptr, string(filename), status)
```
kbarbary: Should all these cases of `bytestring(x)` just be replaced by `x`? `string(x::String) === x` on 0.5 and `bytestring(x::ASCIIString) === x` on 0.4. Was this necessary on 0.3? I'm having a hard time remembering...

jeff-regier:
I still test a few warnings when I run the tests.

```
julia> Pkg.test("FITSIO")
INFO: Testing FITSIO
WARNING: Method definition cfitsio_typecode(Type{Int64}) in module Libcfitsio at /home/jeff/.julia/v0.5/FITSIO/src/libcfitsio.jl:164 overwritten at /home/jeff/.julia/v0.5/FITSIO/src/libcfitsio.jl:164.
WARNING: Base.ASCIIString is deprecated, use String instead.
  likely near /home/jeff/.julia/v0.5/FITSIO/src/libcfitsio.jl:356
WARNING: Base.ASCIIString is deprecated, use String instead.
  likely near /home/jeff/.julia/v0.5/FITSIO/src/libcfitsio.jl:356

WARNING: deprecated syntax "[a=>b for (a,b) in c]".
Use "Dict(a=>b for (a,b) in c)" instead.
WARNING: Base.ASCIIString is deprecated, use String instead.
  likely near /home/jeff/.julia/v0.5/FITSIO/test/runtests.jl:118
All tests passed.
INFO: FITSIO tests passed
```

If you know how to fix these, please let me know.

src/libcfitsio.jl
jeff-regier: No idea...

kbarbary: Looking at the old NEWS.md entries, this was necessary on v0.2 but not on v0.3 and later. So I think we can just get rid of these conversions.

test/runtests.jl
```diff
@@ -117,7 +117,7 @@ write(f, indata; hdutype=ASCIITableHDU)
 # expected output type for each input type.
 expected_type = @compat(Dict(Int16=>Int32, Int32=>Int32,
                      Float32=>Float64, Float64=>Float64,
-                     ASCIIString=>ASCIIString))
+                     ASCIIString=>Compat.ASCIIString))
```
kbarbary: `Compat.ASCIIString=>Compat.ASCIIString`

src/libcfitsio.jl
```diff
@@ -353,8 +353,8 @@ for (a,T,S) in (("ffukys", :ASCIIString, :(Ptr{UInt8})),
                 ("ffukyl", :Bool,        :Cint),
                 ("ffukyj", :Integer,     :Int64))
     @eval begin
-        function fits_update_key(f::FITSFile, key::ASCIIString, value::$T,
-                                 comment::@compat(Union{ASCIIString, Ptr{Void}})=C_NULL)
+        function fits_update_key(f::FITSFile, key::Compat.ASCIIString, value::$T,
```
kbarbary: `$T` is sometimes `ASCIIString`. Change `:ASCIIString` to `:(Compat.ASCIIString)` on L352.

kbarbary:
Did this fix all the warnings?

src/header.jl
```diff
@@ -89,12 +89,12 @@ function fits_try_read_keys{T}(f::FITSFile, ::Type{T}, keys)
     for key in keys
         ccall((:ffgkey, libcfitsio), Cint,
               (Ptr{Void},Ptr{UInt8},Ptr{UInt8},Ptr{UInt8},Ptr{Cint}),
-              f.ptr, bytestring(key), value, C_NULL, status)
+              f.ptr, string(key), value, C_NULL, status)
```
kbarbary: did you miss removing this `string()`?

jeff-regier:
Two warnings left:

```
WARNING: Method definition cfitsio_typecode(Type{Int64}) in module Libcfitsio at /home/jeff/.julia/v0.5/FITSIO/src/libcfitsio.jl:164 overwritten at /home/jeff/.julia/v0.5/FITSIO/src/libcfitsio.jl:164.
```
and 
```
WARNING: deprecated syntax "[a=>b for (a,b) in c]".
Use "Dict(a=>b for (a,b) in c)" instead.
```

jeff-regier:
The second warning is easily fixed by replacing line 114 with
```
        map = Dict(keys[i] => i for i in 1:length(keys))
```
but that breaks compatibility with Julia 0.4. I don't see a macro in Compat.jl that solves the problem

kbarbary:
Could just remove the comprehension: initialize an empty `Dict{String, Int}` and fill it in a loop.

The other warning is because `Clong` is `Int64` on your system. Unrelated to Julia version changes, but maybe there's better method redefinition warnings on 0.5. You could try fixing this with a conditional (only define `cfitsio_typecode(Type{Clong})` if `Clong` is not `Int64`) or it can be done in a separate PR.

jeff-regier:
OK, I think this done now. I'll leave the method definition for another PR.

kbarbary:
Thanks Jeff, looks good.

PR #56, opened 2016-08-10
enable osx on Travis

tkelman:
and use the apt addon so `apt-get install` doesn't try to run on mac

and turn on 0.4 and 0.5 on Travis, since `release` will change meaning soon
0.5 runs the latest RC right now, will change to latest release once final is out


kbarbary:
Good catch, looks good.

PR #57, opened 2017-01-20
Updates for Julia 0.6

andreasnoack:
Mainly the deprecation of the old `Array` constructor. I'm also using `Ref` in `ccall`s whenever possible. Finally, I'm dropping Julia 0.3 support since some of these changes won't work there and I don't think it is relevant to support anymore.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -136,7 +136,7 @@ f = FITS(fname, "w")
 # arre what we expect.
 io = IOBuffer()
 show(io, f)
-s = takebuf_string(io)
+s = String(take!(io))
 @test s[1:6] == "File: "
 @test s[end-7:end] == "No HDUs."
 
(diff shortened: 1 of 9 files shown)


kbarbary:
Thanks for this PR! I definitely agree we can drop Julia 0.3 support.

Looks like `String(take!(io))` in runtests.jl is problematic on Julia 0.4 since it means `AbstractString(take!(io))` there. Maybe we need a custom Julia version check here?

Alternatively, we could also drop Julia 0.4 support.

andreasnoack:
I had forgotten to `import Compat.String`. Should work on 0.4 now as well.

kbarbary:
Great, thanks!

andreasnoack:
Would you mind making a new release? If you don't already know attobot, you should take a look at https://github.com/attobot/attobot. It makes it much simpler.

test/runtests.jl
```diff
@@ -136,7 +137,7 @@ f = FITS(fname, "w")
 # arre what we expect.
 io = IOBuffer()
 show(io, f)
-s = takebuf_string(io)
+s = @compat(String(take!(io)))
```
tkelman: update Compat minimum version for this - 0.9.5

kbarbary: Done.

kbarbary:
JuliaLang/METADATA.jl#7653

Attobot is nice.

PR #58, opened 2017-04-08
Support memory backed files

Keno:
Support read-only memory backed files via `ffomem`. Also clean up a couple of deprecations
while I'm here.

### DIFF
--- a/src/FITSIO.jl
+++ b/src/FITSIO.jl
@@ -80,6 +80,10 @@ type FITS
     mode::AbstractString
     hdus::Dict{Int, HDU}
 
+    # Hold on to memory if backed by a julia buffer
+    memhandle::FITSMemoryHandle
+    hidden::Any
+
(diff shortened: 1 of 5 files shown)


kbarbary:
Thanks! Need a new tagged release for Celeste? Otherwise, will probably wait until 0.6 is out and more deprecation cleanups are done here.

Keno:
No, I think we're ok. This is an option only, so not required to run something.

giordano:
@Keno Should `hidden` have a parameter type `T` and the type `FITS` be parametrized by `T`, rather than having a [field with abstract type](https://docs.julialang.org/en/latest/manual/performance-tips.html#Avoid-fields-with-abstract-type-1)?

Keno:
No, `hidden` is never read, so its type doesn't matter. It's just there to give the GC and edge to the underlying memory to prevent it from being freed.

giordano:
I see, thanks for the explanation.

PR #59, opened 2017-04-10
Clear deprecation warnings in Julia 0.6

giordano:
I don't know what the `@compat` macro is used for in the definitions of `ColumnDef` and `FITSTableScalar`, can be removed?


kbarbary:
Maybe those `@compat`s were left over from supporting 0.3?

giordano:
The package passes test on Julia 0.4.7 on my system without all those `@compat`s.  I removed them and replaced all `convert(T, x)` with `T(x)`.

PR #60, opened 2017-04-10
Fix typo: indicies → indices

giordano:
A typo in variable names may be tolerable, but the word was misspelled also in
an error message, so I fixed it everywhere.


kbarbary:
Thanks. I guess I've been misspelling indices for years. :open_mouth:

PR #62, opened 2017-05-15
Provide FITS method that automatically closes file

giordano:
This is like the `open` method that automatically close the file upon completion: https://docs.julialang.org/en/latest/manual/networking-and-streams.html#Working-with-Files-1


kbarbary:
Looks good to me. Travis failure is unrelated.

PR #64, opened 2017-05-15
Simplify libcfitsio_version

giordano:


giordano:
Uhm, ok, probably this is bad, doesn't work with this example
```julia
julia> VersionNumber(string(3.341))
v"3.341.0"
```

kbarbary:
Yeah this won't work like the original version.

Issue #61, opened 2017-05-15
Can the signature of `Libcfitsio.fits_write_key` be widened?


mweastwood:
I have a `TLONG` that I would like to write to the header using `Libcfitsio.fits_write_key` (the function is mirroring a C routine so I want to use the Libcfitsio interface). At the moment the function only accepts floats and strings (see the link below), but it naively looks like it might have been written with a wider signature in mind.

https://github.com/JuliaAstro/FITSIO.jl/blob/8f5383601c530dcfbce1b9502f82f8f4bcdb54e7/src/libcfitsio.jl#L347-L358

Edit: grammar and typos

giordano:
Out of curiosity, what's a `TLONG` type? How would you suggest to change the signature of the function?

mweastwood:
`TLONG` is just a `Clong` in Julia (can be a 32-bit integer or a 64-bit integer depending on the system). So I'd like `fits_write_key` to at least work with integer `value` as well.

It looks like the function was written under the assumption that `value` might be a `Bool`, but this is prohibited by the function's signature. So I am hoping that the function might still work as written even after simply widening the signature to accept `Bool`, `Int32`, `Int64`, etc.

giordano:
Thanks!  I found out the meaning of `TLONG` in the comment, and right now I was looking to the same inconsistency: a `Bool` `value` is allowed in the definition of `cvalue`, but `value` can't be a `Bool` in the signature of the function.  In other places (like `fits_write_pix`, `fits_read_pix`, `fits_read_subset`, `fits_read_col`, and `fits_write_col`), a generic `T` type is used, without further specification, probably the same can be done here.

kbarbary:
I found the culprit (me in 2014): https://github.com/JuliaAstro/FITSIO.jl/commit/4eb54856a4086cd361e157ca28f72ba3067e1838#diff-7f30490ba5728cf8a2e44da573ed477dL212

Yes, this should definitely be widened! Not sure what happened there.

PR #66, opened 2017-05-18
additional functionality needed for LibHealpix.jl

mweastwood:
I am currently in the processing of rewriting a lot of LibHealpix.jl.
One of its limitations has been that it uses the libchealpix functions
to read and write Healpix maps as FITS images. These functions only work
with Float32 values, so we were frequently discarding precision when
doing I/O with Healpix maps.

To remedy that situation I am replacing this part of the wrapper with
the FITSIO Libcfitsio interface. This makes it easy for me to port the
original C functions into Julia.

This commit adds some missing functionality that I needed to make this
work:

* wrap `fits_read_key_str`
* wrap `fits_read_key_lng`
* wrap `fits_read_keys_lng`
* wrap `fits_write_date`
* widen the signature of `fits_write_key` to accept all `Number`s

ref issue JuliaAstro/FITSIO.jl#61

### DIFF
--- a/src/libcfitsio.jl
+++ b/src/libcfitsio.jl
@@ -103,12 +103,16 @@ export FITSFile,
        fits_read_col,
        fits_read_descript,
        fits_read_keyn,
+       fits_read_key_str,
+       fits_read_key_lng,
+       fits_read_keys_lng,
        fits_read_keyword,
        fits_read_pix,
        fits_read_record,
        fits_read_subset,
        fits_update_key,
        fits_write_col,
+       fits_write_date,
        fits_write_comment,
        fits_write_history,
        fits_write_key,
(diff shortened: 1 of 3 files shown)


giordano:
I was going to ask you some tests, thanks for adding them ;-)  However note that FITSIO.jl supports Julia 0.4, `@testset` isn't available there.

Regarding `fits_write_key`, is Number really necessary? Does Real suffice instead?

giordano:
Looks like something doesn't work on Julia 0.4 (see failures on Travis), and creation of temporary  file doesn't work on Windows (all tests on Windows are failing, see AppVeyor)

mweastwood:
Ok, I *think* I fixed those problems. The CI looks better this time around. Let me know if you want me to squash.

kbarbary:
:+1:  Looks good, thanks!

PR #71, opened 2017-07-19
Trial documentation shift

TestSubjector:
The large work-in-progress PR will come in a while, but I wanted to use this minor PR as a trial run to show how the documentation will look like (some elements are similar to AstroLib.jl).

### DIFF
--- a/src/FITSIO.jl
+++ b/src/FITSIO.jl
@@ -147,6 +147,23 @@ include("header.jl")  # FITSHeader methods
 include("image.jl")  # ImageHDU methods
 include("table.jl")  # TableHDU & ASCIITableHDU methods
 
+"""
+    libcfitsio_version(version=fits_get_version()) -> VersionNumber
+
+### Purpose ###
+
+Return the version of the underlying CFITSIO library
+
+### Example ###
+
+```julia
+julia> using FITSIO
+
+julia> FITSIO.libcfitsio_version()
+v"3.37.0"
+```
+
+"""
 function libcfitsio_version(version=fits_get_version())
     # fits_get_version returns a float. e.g., 3.341f0. We parse that
     # into a proper version number. E.g., 3.341 -> v"3.34.1"
--- a/src/FITSIO.jl
+++ b/src/FITSIO.jl
@@ -148,7 +148,7 @@ include("image.jl")  # ImageHDU methods
 include("table.jl")  # TableHDU & ASCIITableHDU methods
 
 """
-    libcfitsio_version(version=fits_get_version()) -> VersionNumber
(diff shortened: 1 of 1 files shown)


src/FITSIO.jl
```diff
@@ -147,6 +147,23 @@ include("header.jl")  # FITSHeader methods
 include("image.jl")  # ImageHDU methods
 include("table.jl")  # TableHDU & ASCIITableHDU methods
 
+"""
+    libcfitsio_version(version=fits_get_version()) -> VersionNumber
```
giordano: For this particular function, the optional argument may not be documented, it's just a trick to make it possible testing the parsing of version number, I don't expect anybody will use it.

src/FITSIO.jl
```diff
@@ -147,6 +147,23 @@ include("header.jl")  # FITSHeader methods
 include("image.jl")  # ImageHDU methods
 include("table.jl")  # TableHDU & ASCIITableHDU methods
 
+"""
+    libcfitsio_version() -> VersionNumber
+
+### Purpose ###
```
kbarbary: Let's follow the documentation guidelines in the Julia docs and omit this `### Purpose ###` line.

src/FITSIO.jl
```diff
@@ -147,6 +147,23 @@ include("header.jl")  # FITSHeader methods
...
+
+Return the version of the underlying CFITSIO library
+
+### Example ###
+
+```julia
+julia> using FITSIO
```
kbarbary: I think `using FITSIO` can be omitted from examples. Otherwise it would have to be the first line of every single example. Is it necessary for doctests to run correctly?

giordano: Yes, it's necessary, unless documentation is configured to automatically run `using FITSIO` before each snippet.  See https://github.com/JuliaAstro/AstroLib.jl/pull/33#issuecomment-313680563 for the same issue in AstroLib.

TestSubjector: I actually forgot to put `jldoctest` here. But yes, some [setup code](https://juliadocs.github.io/Documenter.jl/v0.8.3/man/doctests.html#Setup-Code-1) would be required for many of the doctests, like that of reading/writing functions etc to run

kbarbary: I'd prefer to omit the line, because I feel it only adds visual noise in reference documentation like this where the user is already aware that they have to import the library. For narrative documentation pages (once we have them) I think it makes sense to add the line for completeness. But... I don't feel very strongly about it.

Yes, I think we shouldn't bother with jldoctest immediately -- it will be tricky for many functions in FITSIO with opening and closing temporary files. We can always add it later.

PR #72, opened 2017-07-22
Drop support for Julia 0.4

giordano:
`Compat.jl` dropped support for Julia 0.4, so it's probably time to do the same here.

Note that `String` type doesn't guarantee there are only ASCII character in the string.  If necessary, that should be tested with `isascii`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ os:
     - linux
     - osx
 julia:
-    - 0.4
     - 0.5
     - 0.6
     - nightly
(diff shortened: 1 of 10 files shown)


kbarbary:
:+1:  Let's open another issue about adding `isascii` checks to all the string-related functions so that we don't forget. This will be a better interface than before: we didn't really want to disallow all the old `UTF8String`s as input: only ones that contained characters outside the ASCII subset.

Looks good to me; feel free to merge.

Issue #75, opened 2017-07-22
Test strings that should contain only ASCII characters with isascii


giordano:
PR #72 dropped support for Julia 0.4, and with it `Compat.ASCIIString` was replaced with `String`.  This type doesn't guarantee there are only ASCII character in the string, there may be places where this should be enforced, `isascii` can be used for this.

PR #77, opened 2017-07-26
Add Documenter to build documentation

TestSubjector:
* Added basic Documenter Setup
* Ported parts of the documentation

TODO
- [x]  Port all the existing documentation to Documenter


src/fits.jl
```diff
@@ -122,7 +122,20 @@ function getindex(f::FITS, name::AbstractString, ver::Int=0)
     return f.hdus[i]
 end
 
+"""
+    close(f::FITS)
 
+### Purpose ###
+
+Close the file. Subsequent attempts to operate on `f will result in an error.
```
giordano: Missing closing backtick after f

src/header.jl
```diff
@@ -193,6 +193,21 @@ end
...
 # Public API
 
+
+"""
+    read_key(hdu, key)
+
+Read the specified key and return a tuple of `(value, comment)`.
+
+The key, can be either the index of the header record (Integer)
+or the header keyword (ASCIIString).
```
giordano: `ASCIIString` has been removed in julia 0.5, use `String` instead. There are also other uses of `ASCIIString`. See also #75.

src/libcfitsio.jl
```diff
@@ -862,6 +1067,28 @@ function fits_read_col{T}(f::FITSFile,
     fits_assert_ok(status[])
 end
 
+"""
+    fits_write_col(f, colnum, firstrow, firstelem, data)
+
+Write some data in one column of a ASCII/binary table.
+
+If there is no room for the elements, new rows will be created. (It is therefore
+useless to call :func:`fits_insert_rows` if you only need to *append* elements
```
kbarbary: Convert `:func:` to `@ref`.

kbarbary:
Looking good so far! I guess me or @giordano will have to set up travis ssh keys, but perhaps that's better done after this PR is merged?

giordano:
There is no rush to do it right now, but I'd do that before merging the PR, so that merge commit readily triggers building of the docs.

kbarbary:
OK, I've set up the GitHub/Travis ssh keys for Documenter.

kbarbary:
How's this going @TestSubjector? If you're short on time, we could do without adding the examples sections for the purpose of this PR.

TestSubjector:
Yeah, this way will work much better for me. I'll work on and make a new PR for the examples documentation, when things become a little less hectic on my side.

kbarbary:
I guess this is ready to merge?

TestSubjector:
All clear from my side 🏁

kbarbary:
Looks like the docs build didn't succeed on Travis. I think we need a `mkdocs.yml` file: https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#The-MkDocs-mkdocs.yml-File-1
and also arguments `deps` and `julia = "0.6"` in the `deploydocs()` function in `docs/make.jl`.

giordano:
Travis on GNU/Linux with nighly Julia is successful, maybe the problem is that the [`src/index.md`](https://github.com/JuliaAstro/FITSIO.jl/blob/c9d5726395781f8c55baee4a7a21e8349012749c/docs/src/index.md) file is actually empty?

kbarbary:
Oh, that's probably not good either.

The travis job can succeed while the doc build fails, because docs are built `after_success`. Looking at the log here: https://travis-ci.org/JuliaAstro/FITSIO.jl/jobs/269730603#L3836

PR #78, opened 2017-08-12
use Homebrew.jl to grab cfitsio on OSX

mweastwood:
I don't have easy access to a Mac so this is untested. Hopefully Travis comes back with a green light.

This *might* fix some of the issues reported in mweastwood/LibHealpix.jl#38, but I think this is a good idea regardless.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
 julia 0.5
 BinDeps 0.3.12
-Compat 0.17.0
+Compat 0.29.0
 @osx Homebrew
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,4 @@
 julia 0.5
 BinDeps 0.3.12
 Compat 0.17.0
+@osx Homebrew
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -1,14 +1,15 @@
 using BinDeps
 using Compat
+using Compat.Sys: iswindows, isapple, isunix
 
 @BinDeps.setup
 
 version = "3370"
 baseurl = "ftp://heasarc.gsfc.nasa.gov/software/fitsio/c/"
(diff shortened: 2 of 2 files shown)


REQUIRE
```diff
@@ -1,3 +1,4 @@
 julia 0.5
 BinDeps 0.3.12
 Compat 0.17.0
+@osx Homebrew
```
kbarbary: Huh, did not know this was legal syntax in REQUIRE!

kbarbary:
Do you know, if an OSX user *doesn't* have homebrew, will build-from-source still be used?

mweastwood:
Homebrew.jl is using is doing its own thing. It doesn't install things globally using the system Homebrew installation so I don't believe having Homebrew is a pre-req for this (Homebrew.jl is though). If this is merged, I think it will always be preferred to the build-from-source method on OSX.

On the plus side though, it means other packages that need libcfitsio (ie. LibHealpix.jl) can use the same installation.

kbarbary:
I see. Seems OK to me, but I'm not on a mac. I guess we can always change it if people have problems using Homebrew.jl. Does anyone with a mac have an opinion?

deps/build.jl
```diff
@@ -62,6 +62,12 @@ if is_windows()
     push!(BinDeps.defaults, BuildProcess)
 end
 
+# OSX
+if is_apple()
```
giordano: Add
```julia
using Compat.Sys: iswindows, isapple, isunix
```
after the `using Compat` line and replace all `is_<os>` with `is<os>`.  You should also rise `Compat` version to 0.29.0 in `REQUIRE`.

mweastwood: updated

PR #80, opened 2017-09-05
Documenter Theme Patch

TestSubjector:
This...should make everything work. Hopefully.


kbarbary:
Nope. You can reproduce this locally by doing
```
julia -e 'cd(Pkg.dir("FITSIO")); include(joinpath("docs", "make.jl"))'
```
which builds the site in `docs/build`, and then doing:
```
cd docs
mkdocs build
```
The problem is that we have native HTML output enabled (`format = :html` in `make.jl`), which means Documenter itself builds an HTML site in `build`. Then it tries to run `mkdocs` which expects to find markdown files in `build` but encounters HTML. See https://juliadocs.github.io/Documenter.jl/stable/man/guide/#Output-formats-1

PR #82, opened 2017-09-18
Do not print newline at the end of show methods

giordano:
Usually `show` methods don't add a newline at the end, as the Julia REPL automatically prints it.

There are no tests, should I add them?

### DIFF
--- a/src/header.jl
+++ b/src/header.jl
@@ -336,14 +336,10 @@ end
 
 # Display the header
 function show(io::IO, hdr::FITSHeader)
-    for i=1:length(hdr)
-        cl = length(hdr.comments[i])
+    n = length(hdr)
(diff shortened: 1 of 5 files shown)


src/header.jl
```diff
@@ -362,7 +363,7 @@ function show(io::IO, hdr::FITSHeader)
                     @printf io " / %s" hdr.comments[i]
                 end
             end
-            print(io, "\n")
+            i != n && print(io, "\n")
```
kbarbary: Looks like there will still be a trailing newline if the last keyword is a HISTORY or COMMENT (L344 and L346).

kbarbary:
Appveyor failure was some transient FTP error on only one Julia version.

PR #83, opened 2017-09-20
Assert strings written to file are ASCII

### LINKED ISSUE #75: Test strings that should contain only ASCII characters with isascii

giordano:
Fix #75.

I think I addressed most cases, but a review is required.

No test yet.

### DIFF
--- a/src/libcfitsio.jl
+++ b/src/libcfitsio.jl
@@ -208,6 +208,9 @@ function fits_assert_ok(status::Cint, filename = nothing)
     end
 end
 
+fits_assert_isascii(str::String) =
+    !isascii(str) && error("FITS file format accepts ASCII strings only")
+
 fits_get_version() = ccall((:ffvers, libcfitsio), Cfloat, (Ptr{Cfloat},), &0.)
 
 # -----------------------------------------------------------------------------
@@ -468,6 +471,8 @@ Write a keyword of the appropriate data type into the CHU.
 """
 function fits_write_key(f::FITSFile, keyname::String,
                         value::Union{Real,String}, comment::String)
+    fits_assert_isascii(keyname)
+    fits_assert_isascii(comment)
     cvalue = isa(value,String) ?  value :
              isa(value,Bool) ? Cint[value] : reinterpret(UInt8, [value])
     status = Ref{Cint}(0)
@@ -485,6 +490,7 @@ function fits_write_date(f::FITSFile)
 end
 
 function fits_write_comment(f::FITSFile, comment::String)
+    fits_assert_isascii(comment)
     status = Ref{Cint}(0)
     ccall((:ffpcom, libcfitsio), Cint, (Ptr{Void}, Ptr{UInt8}, Ref{Cint}),
           f.ptr, comment, status)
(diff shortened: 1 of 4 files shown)


src/libcfitsio.jl
```diff
@@ -58,6 +58,7 @@ module Libcfitsio
 
 export FITSFile,
        FITSMemoryHandle,
+       fits_assert_isascii,
```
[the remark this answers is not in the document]
giordano: This is in Libcfitsio module, it isn't imported when `using FITSIO`. The alternative is to define the function in FITSIO module, but the visibility would be the same.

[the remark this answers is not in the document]
kbarbary: Right, but some users do `using FITSIO.Libcfitsio` and I don't think it should be exported in those cases.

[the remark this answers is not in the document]
giordano: Oh, I didn't know about people `using FITSIO.Libcfitsio`!  I moved the definition to `FITSIO.jl` file.

src/libcfitsio.jl
```diff
@@ -56,6 +56,8 @@ isdefined(Base, :__precompile__) && __precompile__()
 
 module Libcfitsio
 
+import FITSIO: fits_assert_isascii
+
```
kbarbary: This is getting a bit nitpicky, but I think it was better the way you had it, with `fits_assert_isascii` defined in Libcfitsio. Mainly just because we already have some non-exported stuff there that is used by `FITSIO` and follows that pattern (see https://github.com/JuliaAstro/FITSIO.jl/blob/d3d9af24359f3ec90f79866dad2c33736bb50fcd/src/FITSIO.jl#L41-L47). Also it seems cleaner for submodules to not depend on their parent modules, if possible.

src/libcfitsio.jl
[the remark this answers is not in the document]
kbarbary: Yes, I'd rather we just had the high level interface, but putting the low-level wrappers into a submodule was a compromise. See discussion in #27.

src/libcfitsio.jl
giordano: > Also it seems cleaner for submodules to not depend on their parent modules, if possible.

To be honest, I'd have said the other way round.  However I never played much with submodules, so don't know the subtleties of this.

kbarbary: Yeah I might be wrong about that. We can probably agree that it is better to have a one-way rather than bidirectional dependency though!

src/libcfitsio.jl
```diff
@@ -860,6 +871,8 @@ for (a,b) in ((:fits_create_binary_tbl, 2),
         function ($a)(f::FITSFile, numrows::Integer,
                       coldefs::Array{ColumnDef}, extname::String)
```
giordano: I should have mentioned also here that my only doubt was whether the `coldefs` argument should be checked.

src/libcfitsio.jl
giordano: Yes, I agree ;-)

src/libcfitsio.jl
kbarbary: Yeah I do think the name and unit at least should be checked because these get written to the file. I'll add those checks (I'm updating the NEWS file anyway).

PR #86, opened 2017-09-26
Add write_key()

kbarbary:
We had `read_key()` but not `write_key()` for directly manipulating the FITS file header. This adds `write_key()`.

It also clarifies some docstrings about headers.


test/runtests.jl
```diff
@@ -247,36 +256,38 @@ function create_test_file(fname::AbstractString, header::String)
...
+        data = read(f[1])
+        @test data == zeros(10, 10)
+        close(f)
+
+        # test that we can write it back out.
+        f2 = FITS(fname2, "w")
+        write(f2, data; header=hdr)
+        close(f2)
+    finally
+        isfile(fname) && rm(fname)
```
giordano: Now that we require Julia >= 0.5 we can use `rm(fname, force = true)`.

PR #87, opened 2017-10-11
edited possible error in install instructions

paultheastronomer:
Trying to install FITSIO on a clean debian system running Julia Version 0.4.7 I had to remove the .jl extension when adding the package.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -15,7 +15,7 @@ built-in package manager. In a Julia session run the command
 
 ```julia
 julia> Pkg.update()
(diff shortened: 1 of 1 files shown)


giordano:
That's correct, thank you Paul!

However note that the latest versions of `FITSIO.jl` work only with Julia >= 0.5.  Unfortunately, there are currently legal problems upgrading Julia in Debian.  You can either compile Julia yourself (that's what I do in Debian testing) or grab a binary from Julia official website: https://julialang.org/downloads/

PR #89, opened 2017-10-24
Remove no-longer-used sphinx docs

kbarbary:
Everything here should now be in the markdown docs in `docs`.


kbarbary:
Seems like there's something wrong with the OS X nightly Julia build on travis. Merging...

PR #92, opened 2018-01-09
Drop support for unmaintained Julia 0.5

giordano:
There are still some errors on Julia 0.7 that need to be addressed.


kbarbary:
This looks good. Should we merge this now and fix the remaining errors later?

giordano:
Yes, I think so. I hope it'll be easy to fix them with `Compat`, but I didn't check yet.

PR #94, opened 2018-01-12
Add method to get colum names in FITS tables

### LINKED ISSUE #93: Method to get column names in a FITS table
I was surprised that this doesn't already exist, but it's straightforward to add.

Column names are already determined as part of `show(::IO, ::TableHDU)` and `show(::IO, ::ASCIITableHDU)`. One would just have to pull out some of that code into a separate function.

`colnames` might be a good function name for this. DataFrames extends `Base.names` for this purpose, but the definition of `Base.names` is, I think, inconsistent with this usage.

giordano:
Here is my take to close #93.

A couple of comments:
* I'm not sure whether the `colnames` function should be exported
* There are no test about showing an `ASCIITable`.  Why `f[3]` in the "Tables" test set can't be shown?

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -40,6 +40,7 @@ copy_section
 write(::FITS, ::Dict{String})
 write(::FITS, ::Vector{String}, ::Vector)
 read(::TableHDU, ::String)
+FITSIO.colnames(hdu::Union{ASCIITableHDU,TableHDU})
 ```
 
 ## Miscellaneous
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -9,7 +9,7 @@ The interface is inspired by Erin Sheldon's
 
 ## Installation
 
-FITSIO is available for Julia 0.5 and later versions, and can be
+FITSIO is available for Julia 0.6 and later versions, and can be
 installed with [Julia](https://github.com/julialang/julia.jl)'s
 built-in package manager. In a Julia session run the command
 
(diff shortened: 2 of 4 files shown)


kbarbary:
Looks good!

> There are no test about showing an ASCIITable. Why f[3] in the "Tables" test set can't be shown?

The lack of a test was an oversight or laziness on my part. Do you mean you tried to show it and an error resulted? Here's a test that should pass:

```julia
    # test show/repr on ASCIITableHDU by checking that a couple lines are what we expect
    lines = split(repr(f[3]), "\n")
    @test lines[4] == "Rows: 20"
    @test lines[6] == "         col3  Float64  E26.17  "
```

> I'm not sure whether the colnames function should be exported

Not having a strong opinion, I'd err by being conservative and *not* exporting it. We can always add the export later but it's harder to take it away.

giordano:
> Do you mean you tried to show it and an error resulted?

Exactly.  I tried to replicate the `colnames` test for `ASCIITableHDU`, but accessing `f[3]` gave me an error about the `CHDU`.  I also tried to just print `f[3]`, but it resulted in the same error, of course.

However, I can't reproduce the error anymore, I have no idea what was going on yesterday :-/

PR updated, with tests for `colnames` and `show` on `ASCIITableHDU`.

> I'd err by being conservative and not exporting it. We can always add the export later but it's harder to take it away.

Completely agree, that's why I didn't export ;-)

kbarbary:
Nice! If you want to add colnames to the API docs as part of this, that would be good. Just so we don't forget.

giordano:
Done!  I've also update a couple of other docstrings still mentioning `ASCIIString`.

PR #96, opened 2018-01-30
Switch binary dependency to use BinDeps2

Keno:
The builder repository is at https://github.com/Keno/CFITSIOBuilder, but should probably be transferred to JuliaAstro. In the process I had to bump the cfitsio version to 3420 from 3370
since that included some build system fixes, though I could probably backport those to 3370
if required.

BinDeps2 is somewhat experimental still, but if CI goes through, this should be usable.


giordano:
> if CI goes through, this should be usable

Something went wrong everywhere.  Apparently, Travis + GNU/Linux + Julia 0.6 was able to get the dependency, but then the first test fails with
```
  ccall: could not find function ffinit in library /home/travis/.julia/v0.6/FITSIO/deps/usr/lib/libcfitsio.so
```

Keno:
Yeah, might either be a problem with the library itself or might have something to do with the version bump. I'll investigate.

giordano:
@Keno see #97

mweastwood:
Thanks for your help getting CFITSIOBuilder working, Keno.

This PR is redundant now that #97 has been merged.

Issue #49, opened 2016-03-21
FITSIO doesn't seem to check if CFITSIO is already installed.


rcthomas:
It always seems to try to download and build CFITSIO.

StefanKarpinski:
You probably want to provide the output of `versioninfo()` as well.

rcthomas:
Thanks, sorry!

    julia> versioninfo()
    Julia Version 0.5.0-dev+2918
    Commit 1eacbbb* (2016-02-28 16:56 UTC)
    Platform Info:
      System: Linux (x86_64-suse-linux)
      CPU: Intel(R) Xeon(R) CPU E5-2698 v3 @ 2.30GHz
      WORD_SIZE: 64
      BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Haswell)
      LAPACK: libopenblas64_
      LIBM: libopenlibm
      LLVM: libLLVM-3.7.1

StefanKarpinski:
Thanks. How did you install CFITSIO? System package manager or manually from source?

kbarbary:
Hmm... I thought that BinDeps (which is how FITSIO satisfies dependencies) checks for an existing library before  trying to satisfy the dependency in another way, but I might be wrong about that.

StefanKarpinski:
That's why I was asking about how CFITSIO was installed. If it's installed via system package manager, then that version should get picked up by BinDeps if it's working right. If it was installed manually from source, it's pretty hard to make the package reliably use that version.

rcthomas:
It's built from source.   Does BinDeps look at pkg-config?

mweastwood:
Closed by #97. FITSIO.jl now uses BinDeps2 to download pre-compiled binaries.

PR #97, opened 2018-04-20
switch to using BinaryBuilder/BinaryProvider

mweastwood:
This should replace #96.

I've tested this locally (Linux x86_64, and things seem to work fine here at least). I believe the additional version bump relative to #96 resolves the issues that were seen there. We're also using more recent versions of BinaryBuilder and BinaryProvider.

Here we are affected by staticfloat/ObjectFile.jl#7 because cfitsio builds a fat macho on OSX, which isn't currently supported by this workflow. I believe recent changes to BinaryBuilder make it possible to work around this problem, but the usual auditing of the binary is not performed.

I have this hosted on my own Github account for now, but it should definitely be transferred to JuliaAstro at some point.

Link to the builder repo: https://github.com/mweastwood/CFITSIOBuilder


mweastwood:
Hmm, it looks like there is a problem on Windows. I have a Windows PC at home, so I'll be able to take a look within the next week or so (out of town at the moment).

My initial guess is that it is maybe related to this warning that I get during the build process:
```
Warning: lib/libcfitsio.dll should be in `bin`
```

mweastwood:
Alright, perfect. This is now passing on all platforms.

mweastwood:
I have transferred CFITSIOBuilder to JuliaAstro. This is basically ready to be merged now.

helgee:
LGTM

mweastwood:
If there are no objections I'm going to merge this tomorrow.

This will close #49 because we no longer look for the system CFITSIO under any circumstances. On the plus side we have to do infinitely less worrying about all the different possible system configurations, and we know exactly which version of CFITSIO all of our users are using.

PR #99, opened 2018-04-27
Make fits_get_colnum case sensitive by default

### LINKED ISSUE #98: fits_get_colnum should be case-sensitive
It seems that fits_get_colnum is set-up to be always case insensitive.
I know a fair number of astronomical catalogs that use case-sensitive column names.
The Gaia DR2 catalog (as offered on [CDS ViZieR](http://vizier.u-strasbg.fr/viz-bin/VizieR-3?-source=I/345/gaia2)) released yesterday is one of them (e.g. the b_Teff and B_Teff columns).
It would be great to make it case sensitive by default.
thanks

giordano:
Add an optional keyword to select the behavior.  Fix #98.

### DIFF
--- a/src/libcfitsio.jl
+++ b/src/libcfitsio.jl
@@ -929,7 +929,7 @@ for (a,b,T) in ((:fits_get_num_cols,  "ffgncl",  :Cint),
     end
 end
 
-function fits_get_colnum(f::FITSFile, tmplt::String)
+function fits_get_colnum(f::FITSFile, tmplt::String; case_sensitive::Bool=true)
     result = Ref{Cint}(0)
     status = Ref{Cint}(0)
 
@@ -937,7 +937,7 @@ function fits_get_colnum(f::FITSFile, tmplt::String)
     #                                                1 = case-sensitive
     ccall(("ffgcno", libcfitsio), Cint,
           (Ptr{Void}, Cint, Ptr{UInt8}, Ref{Cint}, Ref{Cint}),
-          f.ptr, 0, tmplt, result, status)
+          f.ptr, case_sensitive, tmplt, result, status)
     fits_assert_ok(status[])
     return result[]
 end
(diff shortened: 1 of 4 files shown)


src/libcfitsio.jl
```diff
@@ -929,15 +929,15 @@ for (a,b,T) in ((:fits_get_num_cols,  "ffgncl",  :Cint),
...
+function fits_get_colnum(f::FITSFile, tmplt::String; case_sensitive::Bool=true)
     result = Ref{Cint}(0)
     status = Ref{Cint}(0)
 
     # Second argument is case-sensitivity of search: 0 = case-insensitive
     #                                                1 = case-sensitive
     ccall(("ffgcno", libcfitsio), Cint,
           (Ptr{Void}, Cint, Ptr{UInt8}, Ref{Cint}, Ref{Cint}),
-          f.ptr, 0, tmplt, result, status)
+          f.ptr, Int(case_sensitive), tmplt, result, status)
```
kbarbary: `Cint` for clarity, or maybe just leave off the conversion and ccall will convert it automatically.

src/table.jl
```diff
@@ -444,12 +444,12 @@ function fits_read_var_col(f::FITSFile, colnum::Integer, data::Vector{String})
 end
 
 # Read a table column
-function read(hdu::ASCIITableHDU, colname::String)
+function read(hdu::ASCIITableHDU, colname::String, case_sensitive::Bool=true)
```
kbarbary: `,` -> `;` ?

src/libcfitsio.jl
giordano: Good point, thanks.

src/table.jl
giordano: Yes, of course!  Fixed, thanks

kbarbary:
LGTM.

PR #101, opened 2018-07-05
[WIP] Fixes for Julia v0.7

mweastwood:
I ran FemtoCleaner locally to get the first set of fixes, and manually fixed the remaining errors and deprecations that came up when running the tests. There is still a pair of deprecations that seem to come from the `deps.jl` file generated by `BinaryProvider`.

The biggest change here is replacing the use of `Nullable` in the keyword parsing routines. I believe these functions take inspiration from `Base.tryparse`, which used to return a `Nullable`. The return type for `Base.tryparse` is now `Union{T, nothing}`, so I've adapted our interface to match. One tricky part is that we need to distinguish between "I was able to parse the keyword, and the result is `nothing`" and "I wasn't able to parse the keyword, so I am returning `nothing` as a sentinal". Apparently in the former case, we should use `Some(nothing)`, so that is what I have implemented.

I'm leaving this as a WIP because I'd like to check that `LibHealpix.jl` is happy with these changes as well before merging. I also do not yet know if I need to generate a `Project.toml` file for us to be `Pkg3` compatible.


mweastwood:
I will also note that there is a new version of `cfitsio`, so it would also be nice to bump that dependency.

helgee:
For packages already in METADATA there is no Project.toml file needed for now because they will be added upstream at some point.

kbarbary:
Yep, you're right about keyword parsing being inspired by `Base.tryparse`. The changes you made there look good to me!

PR #106, opened 2018-09-14
Issue deprecation warning for read(::TableHDU, ::String) with no keyword

giordano:
Use a `try`-`catch` block to issue the warnings.  Ref. #98.

This is done only for `read` because it's the user-visible function.  I don't see how we can issue the warning only once if it's also inside `fits_get_colnum`.

I'd also like to add a couple of tests to make sure that the warning is emitted in the correct case, but I can't make `@test_warn` and `@test_nowarn` work.  If anyone has a clue about how to do this, please share!


kbarbary:
Expertly done! Nice job.

giordano:
Indeed I'm getting some experience with deprecation warnings :sweat_smile:

PR #109, opened 2019-01-13
Add script for Cirrus CI for testing on FreeBSD

giordano:
Following the example of [`SpecialFunctions.jl`](https://github.com/JuliaMath/SpecialFunctions.jl), I've prepared a script for testing the package on FreeBSD by using Cirrus CI.  This can be particularly useful for packages with binary dependencies.

For the time being, I've set up Cirrus CI only for this repository under JuliaAstro, but if we find this service useful we can automatically enable it for all repositories, current and future.


giordano:
We are not currently building FITSIO for FreeBSD :man_facepalming:

giordano:
Actually [CFITSIOBuilder](https://github.com/JuliaAstro/CFITSIOBuilder) already provided a build for FreeBSD, but the old `build.jl` script in this repository doesn't have a platform entry for FreeBSD.  I've updated the `build.jl` script in #110 and rebased this PR on it, so now we have support for FreeBSD too, and it works!

Issue #116, opened 2019-06-15
Modifying an image in a FITS file


evgenyneu:
Is there a way to **modify** image data in a .fits file without touching anything else?

I have a .fits file that contains image data as well as several extensions (tables). I want to replace the image data while keeping everything else (headers/extensions etc.) unchanged. Is there a way to do it?

Right now, I'm creating a new file and copying all the extensions to it:

```Julia
hdul_data = FITS("input.fits")

# Load the image data
image_data = read(hdul_data[1])

# Modify data
image_data[10, 10] = 42

# Save the data array into the output fits file
data_out = FITS("output.fits", "w");

for hdu in hdul_data
  if typeof(hdu)==FITSIO.ImageHDU
    # Write the image to output file
    write(data_out, image_data; header=header_data)
  elseif typeof(hdu)==FITSIO.TableHDU
    # Copy the table into the output file
    table_data = Dict{String,Array}()

    for colname in FITSIO.colnames(hdu)
      col_data = read(hdu, colname)
      table_data[colname] = col_data
    end

    header = read_header(hdu)
    ext_name = header["EXTNAME"]
    write(data_out, table_data; hdutype=FITSIO.TableHDU, header=header, name=ext_name)
  end
end
```

This is not ideal, because I'm not sure if the new file will be exactly the same as the old one (except for the changed image). I would prefer to just modify the image data. These are example of how this could be done in Python and Fortran. 

Python example:

```Python
from astropy.io import fits

hdul_data = fits.open("data.fits")

# Load the image data
image_data = hdul_data["PRIMARY"].data

# Change image data
image_data[5,5]=123

# Save the data array into the output fits file
hdul_data.writeto("output.fits")
```

Fortran example (cfitsio):
```Fortran
! Get an unused Logical Unit Number to use to open the FITS file.
call ftgiou(unit_data,status)

! Open the FITS file, with read-only access.
readwrite=0
call ftopen(unit_data,filename_data,readwrite,blocksize,status)

! Load the image data
group=1
nullval=0
call ftg2di(unit_data, group, nullval, max_x, naxes_data(1), naxes_data(2), image_data, anynull, status)

! Modify the image
image_data(10,10) = 42

! Create the output fits file
! -------------
 
! Get an unused Logical Unit Number to use to create the FITS file.
call ftgiou(unit_output, status)

! Create the output file
blocksize=1
call ftinit(unit_output, filename_output, blocksize, status)

! Copy the entire data file to the output file
[...]

PR #117, opened 2019-06-28
Write data to an existing image HDU

evgenyneu:
Added a feature discussed in #116.

### DIFF
--- a/src/image.jl
+++ b/src/image.jl
@@ -179,6 +179,17 @@ function write(f::FITS, data::Array{T};
     nothing
 end
 
+"""
+    write(hdu::ImageHDU, data::Array)
+
+Write data to an existing image HDU.
+"""
+function write(hdu::ImageHDU, data::Array{T}) where T
+    fits_assert_open(hdu.fitsfile)
+    fits_write_pix(hdu.fitsfile, ones(Int, length(size(data))), length(data), data)
+    nothing
+end
+
 # Copy a rectangular section of an image and write it to a new FITS
 # primary image or image extension. The new image HDU is appended to
 # the end of the output file; all the keywords in the input image will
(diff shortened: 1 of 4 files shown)


src/image.jl
```diff
@@ -179,6 +179,17 @@ function write(f::FITS, data::Array{T};
     nothing
 end
 
+"""
+    write(hdu::ImageHDU, data::Array)
+
+Write data to an existing image HDU.
+"""
+function write(hdu::ImageHDU, data::Array{T}) where T
+    fits_assert_open(hdu.fitsfile)
```
kbarbary: One thing missing here: moving to the right HDU: `fits_movabs_hdu(hdu.fitsfile, hdu.ext)`

src/image.jl
```diff
@@ -179,6 +179,17 @@ function write(f::FITS, data::Array{T};
...
+    fits_write_pix(hdu.fitsfile, ones(Int, length(size(data))), length(data), data)
```
kbarbary: You might want to add a check that the dimensions of `data` match the dimensions of the data already in the HDU. I think you can use `sz = fits_get_img_size(hdu.fitsfile)` for this.

src/image.jl
```diff
@@ -179,6 +179,27 @@ function write(f::FITS, data::Array{T};
...
+
+    # Ensure sizes are equal
+    hdu_size = fits_get_img_size(hdu.fitsfile)
+    data_size = collect(size(data))
+
+    if hdu_size != data_size
+        error("size of HDU $(hdu_size) not equal to size of data $(data_size).")
+    end
+
+    fits_movabs_hdu(hdu.fitsfile, hdu.ext)
```
kbarbary: This (`fits_movabs_hdu`) actually has to come right after the `fits_assert_open` call. The `hdu.fitsfile` C object internally keeps track of which HDU it is on (the "current" HDU). It could be on *any* HDU at the start of this function. So we need to move to the right HDU before checking the image sizes.

evgenyneu: I'm so dumb :)

kbarbary:
Looks good to me! I'll let @giordano merge...

giordano:
Thank you so much, very nice job!

PR #128, opened 2020-03-22
Compact syntax for getting, setting header values and comments

sefffal:
Hi all,
Thanks for this package.

Here is an idea and implementation for a compact syntax for getting and setting header values and comments at the same time:
```julia
# Make a header to use in demo
header = FITSHeader(["key1"], [1], ["comment 1"])

# Get a value and comment using :
value, comment = header["key1", :]

# Set a new value and comment
header["key2",:] = 2.0, "the second comment"
```

This was done by overriding `getindex` and `setindex!` for the colon argument.
If this is interesting, I would be happy to add tests etc.

Thanks for reading!

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "FITSIO"
 uuid = "525bcba6-941b-5504-bd06-fd0dc1a4d2eb"
-version = "0.16.9"
+version = "0.16.10"
 
 [deps]
 CFITSIO = "3b1b4be9-1499-4b22-8d78-7db3344d1961"


kbarbary:
Cool. I think a compact syntax for this would be great. Colon seems a little weird, since I'd expect `header["key", :]` to return a 1-d array, in analogy with other data structures (like 2-d Arrays). I was wondering if using a special symbol value like `header["key", :comment]` would be more explicit?

mileslucas:
Or perhaps something like `header[“key”, !]` if we want to really differentiate it from slice indexing (or any other small symbol  function like ^, *, +, <, etc) without writing out “:comment”.

emmt:
Why not passing a tuple `(val, com)` with the value and comment instead of just `val` when you just want to change the value?  This is what I have done in [EasyFITS.jl](https://github.com/emmt/EasyFITS.jl) (an attempt to simplify dealing with FITS files).

mileslucas:
I think that creates an asymmetry in design, because that would only work with `setindex!’, we still need a way to get the comment out using some compact notation

emmt:
OK then back to the previous suggestion, why not with an additional `/` after the key as a reminder that this symbol is the value-comment separator in FITS header?

sefffal:
Good points about the colon, it is a bit odd. I like the `/` suggestion since it looks like a FITS comment.
Would we use it for directly accessing the comment (not a value and comment tuple)? 

e.g.
```julia
header["KEY"] = 1
header["KEY", /] = "comment"
```

PR #138, opened 2020-07-01
Switch to CFITSIO.jl

mileslucas:


mileslucas:
drone errors are due to lack of Julia 1.4 image

giordano:
Yeah, I know... :disappointed: Julia 1.3 might work for 32-bit arm though

mileslucas:
What still needs to be done for this?

mileslucas:
@giordano oops, I think I needed to remove the `deps` folder. `build.jl` is being executed on a fresh install and failing because `BinaryProvider` isn't in the Project.toml. We shouldn't need the build script at all right? Testing locally right now.

giordano:
Uh, yeah, my bad that didn't notice, but weird that this wasn't caught by CI :confused:

mileslucas:
[it looks like it was "caught" by CI](https://travis-ci.org/github/JuliaAstro/FITSIO.jl/jobs/714256018#L251), it just didn't break any of the tests because we didn't need any of the products from the build script. As it stands, this is just an installation inconvenience- I can confirm that even with the error FITSIO loads and works fine.

giordano:
I see.

PR #140, opened 2020-08-03
Added method to get default header for `AbstractArray`

siddharthlal25:
Added this method to get default header of an `AbstractArray`, it is being used in `CCDReduction.jl` to create `CCDData` directly from an `AbstractArray`. E. g.
```
CCDData(data::AbstractArray) = CCDData(data, get_default_header(data))
```

### DIFF
--- a/src/header.jl
+++ b/src/header.jl
@@ -424,11 +424,11 @@ function show(io::IO, hdr::FITSHeader)
 end
 
 """
-    get_default_header(data::AbstractArray)
+    default_header(data::AbstractArray)
 
 Creates a default header for the given array with the `SIMPLE`, `BITPIX`, `NAXIS`, `NAXIS*`, and `EXTEND` entries.
 """
-function get_default_header(data::AbstractArray{T}) where T <: Number
+function default_header(data::AbstractArray{T}) where T <: Number
     # assigning keys
     hdu_keys = ["SIMPLE",
                 "BITPIX",
--- a/src/header.jl
+++ b/src/header.jl
@@ -433,12 +433,12 @@ function get_default_header(data::AbstractArray{T}) where T <: Number
     hdu_keys = ["SIMPLE",
                 "BITPIX",
                 "NAXIS",
-                [string(Symbol("NAXIS", i)) for i in 1:ndims(data)]...,
+                ("NAXIS$i" for i in 1:ndims(data))...,
                 "EXTEND"]
 
     # assiging values
     hdu_values = [true,                                           # SIMPLE
-                  FITSIO.Libcfitsio.bitpix_from_type(T),          # BITPIX
+                  Libcfitsio.bitpix_from_type(T),          # BITPIX
                   ndims(data),                                    # NAXIS
                   reverse(size(data))...,                         # size of each axis
                   true]                                           # EXTEND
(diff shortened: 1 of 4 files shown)


src/header.jl
```diff
@@ -422,3 +422,33 @@ function show(io::IO, hdr::FITSHeader)
         i != n && println(io)
     end
 end
+
+"""
+    get_default_header(data::AbstractArray)
+
+Creates a default header from an `AbstractArray` with fields being `SIMPLE`, `BITPIX`, `NAXIS`, axis lengths and `EXTEND`.
```
mileslucas, line 429:
```suggestion
Creates a default header for the given array with the `SIMPLE`, `BITPIX`, `NAXIS`, `NAXIS*`, and `EXTEND` entries.
```

src/header.jl
```diff
@@ -422,3 +422,33 @@ function show(io::IO, hdr::FITSHeader)
...
+"""
+function get_default_header(data::AbstractArray{T}) where T <: Number
+    # assigning keys
+    hdu_keys = ["SIMPLE",
+                "BITPIX",
+                "NAXIS",
+                [string(Symbol("NAXIS", i)) for i in 1:ndims(data)]...,
```
mileslucas, line 436: `["NAXIS$i" for i in 1:ndims(data)]...`

is there a way to accomplish this withou allocating this intermediate array? (does this even allocate the intermediate array?)

src/header.jl
```diff
@@ -422,3 +422,33 @@ function show(io::IO, hdr::FITSHeader)
...
+                  FITSIO.Libcfitsio.bitpix_from_type(T),          # BITPIX
+                  ndims(data),                                    # NAXIS
+                  reverse(size(data))...,                         # size of each axis
+                  true]                                           # EXTEND
+
+    # assigning comments
+    comments = ["file does conform to FITS standard",                                   # comment for SIMPLE
+                "number of bits per data pixel",                                        # comment for BITPIX
+                "number of data axes",                                                  # comment for NAXIS
+                [string(Symbol("length of data axis ", i)) for i in 1:ndims(data)]...,  # comments for axis length
```
mileslucas, line 450: same comment as above

src/header.jl
siddharthlal25: There were around 60 allocations on this version (`@benchmark get_default_header(randn(512, 512))`), the newer version comes down to 48.

kbarbary:
`get_...` gives me the impression that the thing you're getting is already an attribute or property of the argument, but in this case, you're *creating* a header based on the image. I suggest just `default_header(data)`.

src/header.jl
```diff
@@ -422,3 +422,33 @@ function show(io::IO, hdr::FITSHeader)
...
+    # assigning keys
+    hdu_keys = ["SIMPLE",
+                "BITPIX",
+                "NAXIS",
+                ("NAXIS$i" for i in 1:ndims(data))...,
+                "EXTEND"]
+
+    # assiging values
+    hdu_values = [true,                                           # SIMPLE
+                  Libcfitsio.bitpix_from_type(T),                 # BITPIX
```
giordano, line 441: This module is deprecated now.  I guess this should be something like
```suggestion
                  CFITSIO.bitpix_from_type(T),                 # BITPIX
```

PR #142, opened 2020-08-03
import required functions from CFITSIO

aplavin:


giordano:
Adding tests to make sure we don't break this again would be great :wink:

aplavin:
Currently I don't see any tests that involve named HDUs, and I don't know how to create fits-file with such HDUs from scratch.

PR #179, opened 2021-12-25
Delete key in FITSHeader

jbrinchmann:
Hi, 

I need to delete keywords in FITSHeaders relatively frequently so I implemented a deletekey function which perhaps is of interest. This PR has that implemented as well as adding it to exports and putting in a simple test case.

### DIFF
--- a/src/header.jl
+++ b/src/header.jl
@@ -263,6 +263,9 @@ function write_key(hdu::HDU, key::String,
 end
 
 
+
+
+
 """
     read_header(filename::AbstractString, hduindex = 1) -> FITSHeader
 
@@ -382,6 +385,37 @@ end
 function setindex!(hdr::FITSHeader, value::Any, i::Integer)
     hdr.values[i] = value
 end
+
+
+"""
+    delete!(hdr::FITSHeader, key::String)
+
+Delete a key in a FITS header. 
+"""
+function deletekey!(hdr::FITSHeader, key::String)
+    if (haskey(hdr, key))
+        index = hdr.map[key]
+
+        # Delete the entries
+        deleteat!(hdr.values, index)
+        deleteat!(hdr.keys, index)
+        deleteat!(hdr.comments, index)
+
+        # Then re-create the map array
+        map = Dict{String, Int}()
+        for i in 1:length(hdr.keys)
+            map[hdr.keys[i]] = i
+        end
+        hdr.map = map
+
+    else
+        println("No key $key found!")
+    end
+
+    return nothing
+end
+
+
 
 # Comments
 """
--- a/src/header.jl
+++ b/src/header.jl
@@ -393,24 +393,20 @@ end
 Delete a key in a FITS header. 
 """
 function delete!(hdr::FITSHeader, key::String)
-    if (haskey(hdr, key))
-        index = hdr.map[key]
 
-        # Delete the entries
-        deleteat!(hdr.values, index)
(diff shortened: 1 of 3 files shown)


src/header.jl
```diff
@@ -383,6 +386,37 @@ function setindex!(hdr::FITSHeader, value::Any, i::Integer)
     hdr.values[i] = value
 end
 
+
+"""
+    delete!(hdr::FITSHeader, key::String)
+
+Delete a key in a FITS header. 
+"""
+function deletekey!(hdr::FITSHeader, key::String)
```
mileslucas, line 395: Why not override `Base.delete!`?

jbrinchmann: Good question! It should have been - indeed the documentation had delete! in the text.  I have committed a change to override delete! which also matches better FITSIO's other header handling.

src/header.jl
```diff
@@ -383,6 +386,37 @@ function setindex!(hdr::FITSHeader, value::Any, i::Integer)
...
+        # Delete the entries
+        deleteat!(hdr.values, index)
+        deleteat!(hdr.keys, index)
+        deleteat!(hdr.comments, index)
+
+        # Then re-create the map array
+        map = Dict{String, Int}()
+        for i in 1:length(hdr.keys)
+            map[hdr.keys[i]] = i
+        end
```
mileslucas, line 408: Could this work as
```suggestion
        map = Dict(zip(hdr.keys, 1:length(hdr.keys)))
```

src/header.jl
```diff
@@ -383,6 +386,37 @@ function setindex!(hdr::FITSHeader, value::Any, i::Integer)
...
+        hdr.map = map
+
+    else
+        println("No key $key found!")
+    end
```
mileslucas, line 413: I would recommend throwing an error instead of just printing a line, especially to stay in line with what you would expect from a Base Dictionary, since we're using that API.
```suggestion
    haskey(hdr, key) || throw(KeyError(key))
    
    index = hdr.map[key]

    # Delete the entries
    deleteat!(hdr.values, index)
    deleteat!(hdr.keys, index)
    deleteat!(hdr.comments, index)

    # Then re-create the map array
    map = Dict{String, Int}()
    for i in 1:length(hdr.keys)
        map[hdr.keys[i]] = i
    end
    hdr.map = map

```

test/runtests.jl
```diff
@@ -738,6 +738,11 @@ HISTORY this is a history"""
             # Test that show() works and that the beginning of output is what we expect.
             @test repr(f)[1:6] == "File: "
 
+            # Test the deletion of a key
+            dhdr = deepcopy(inhdr)
+            delete!(dhdr, "FLTKEY")
+            @test !haskey(dhdr, "FLTKEY")
+
```
mileslucas, line 745:
```suggestion
@test_throws KeyError delete!(dhdr, "aaabbbbccccdddd")
```
make sure we throw an error, as expected

mileslucas (changes requested):

src/header.jl
jbrinchmann: Yes, that certainly works fine and given the simplicity of the code that is probably tidier.

test/runtests.jl
jbrinchmann: Implemented - all good suggestions, thanks

mileslucas:
This looks good, thanks for contributing!

PR #184, opened 2022-07-31
properly define get() instead of getkey()

aplavin:
the current getkey() behavior should actually be get(): getkey in julia returns the key, while get returns the value


src/header.jl
```diff
@@ -364,8 +364,8 @@ Array of values in header of HDU (not a copy).
 """
 values(hdr::FITSHeader) = hdr.values
 
-getkey(hdr::FITSHeader, key::String, def) =
-    (haskey(hdr, key) ? getindex(hdr, key) : def)
+@deprecate getkey(hdr::FITSHeader, key::String, def) get(hdr, key, def)
+get(hdr::FITSHeader, key::String, def) = haskey(hdr, key) ? hdr[key] : def
```
mileslucas, line 368:
```suggestion
getkey(hdr::FITSHeader, key::String, default) = haskey(hdr, key) ? key : default
get(hdr::FITSHeader, key::String, default) = haskey(hdr, key) ? hdr[key] : default
```

test/runtests.jl
```diff
@@ -690,8 +690,8 @@ HISTORY this is a history"""
 
             # Test reading possibly missing keyword
             @test_throws KeyError inhdr["BADKEY"]
-            @test getkey(inhdr, "BADKEY", nothing) === nothing
-            @test getkey(inhdr, "INTKEY", nothing) == inhdr["INTKEY"]
+            @test get(inhdr, "BADKEY", nothing) === nothing
+            @test get(inhdr, "INTKEY", nothing) == inhdr["INTKEY"]
```
mileslucas, line 695:
```suggestion
            @test getkey(inhdr, "BADKEY", nothing) === nothing
            @test getkey(inhdr, "INTKEY", nothing) == "INTKEY"
```

mileslucas (changes requested):
Thanks for catching this @aplavin! However, I think rather than approaching this as a deprecation we should approach it as a bugfix. By that I mean, making sure both `get` and `getkey` are properly defined. I've added some suggested changes.

@giordano do you think a bugfix like this should provoke a minor release or a patch release?

mileslucas:
I have no qualms with a breaking release, I say go for it.

mileslucas:
That sounds like a good idea. Thanks for the help!

PR #190, opened 2023-05-16
Deletefix

jbrinchmann:
There is a bug in header.jl in the delete! function where the map was not updated in the FITSHeader which could cause problems when multiple keys are deleted in a header. This fixes that and also adds a test in runtests.jl to verify that deletion of two keys in sequence works.


mileslucas (approved):
Thanks for opening up a PR for this. It looks good to me, appreciate the test case. Will merge and push a patch version shortly.

PR #240, opened 2026-03-20
Fix WCSExt to properly parse FITS keyword types using `parse_header_val`

### LINKED ISSUE #239: FITS Header type conversions "nonstandard"
When doing the below.

```
wcs_all = WCSTransform(2;
            cdelt = cdel_list[1],
            ctype = ["RA---TAN", "DEC--TAN"],
            crpix = [1.0, 1.0],
            crval = [minra, mindec],
)
wcs_header = FITSHeader(wcs_all)
```

Writing this to disk, and then reading it back, I find that many keywords that should be Ints are Ints (like NAXIS, NAXIS1, etc.). However, many of the other non structural keywords that should be Ints and Floats are ONLY strings, like CRVAL1, CRVAL2, WCSAXES. My astropy friends tell me it is "noncompliant" with standards that they are being read in as strings. Thoughts?

cgarling:
Fixes #239. @andrew-saydjari could you demo this with your workflow to verify it fixes the problem you noticed?

### DIFF
--- a/ext/WCSExt.jl
+++ b/ext/WCSExt.jl
@@ -26,12 +26,12 @@ julia> FITSHeader(wcs)
 WCSAXES =                    2 / Number of coordinate axes
 CRPIX1  =              -234.75 / Pixel coordinate of reference point
 CRPIX2  =               8.3393 / Pixel coordinate of reference point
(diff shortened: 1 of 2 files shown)


andrew-saydjari:
Sorry for my delay on this. Confirmed. Thanks so much!

